PR #35, opened 2019-08-21
Latex

ratnania:
other improvements will be done in the future

### DIFF
--- a/sympde/printing/latex.py
+++ b/sympde/printing/latex.py
@@ -4,6 +4,7 @@
 from sympy.core import Symbol
 from sympy import Mul, Tuple
 from sympy.printing.latex import LatexPrinter as LatexPrinterSympy
+from sympy.printing.latex import translate
 
 #from sympde.expr import BilinearForm, LinearForm, Integral, FormCall
 from sympde.calculus import Dot, Inner, Cross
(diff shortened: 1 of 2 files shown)


sympde/printing/latex.py
```diff
@@ -76,60 +141,89 @@ def _print_Cross(self, expr):
...
+            dx = r'~d\mathbf{x}'
+
+        return '{integral} {expr} {dx}'.format(integral=integral,
+                                               expr=expr,
+                                               dx=dx)
+    # ...
+
+    # ...
+    def _print_BoundaryIntegral(self, expr):
+        domain = self._print(expr.domain)
```
yguclu: I have just noticed that we use the same word, `domain`, to identify the domain of integration as well as the domain of our PDE, i.e. `\Omega`. Is there a way to avoid this confusion?

yguclu:
When running the tests we get lots of deprecation warnings, most of which are due to the presence of invalid escape sequences in Python strings. This can be avoided by using **raw strings**, which are obtained by prepending the character `r` to the strings.

sympde/printing/latex.py
ratnania: It is indeed the case when we have Omega (I added a call to translate from sympy). However, for the moment, if we have a variable such as Omega_1, it is not translated. This should be improved in the future.

ratnania:
> When running the tests we get lots of deprecation warnings, most of which are due to the presence of invalid escape sequences in Python strings. This can be avoided by using **raw strings**, which are obtained by prepending the character `r` to the strings.

There were 5 or 6 strings where the **r** character was missing. Fixed now.

PR #40, opened 2019-08-22
Devel 30

ratnania:


sympde/expr/tests/test_expr_2d.py
```diff
@@ -1723,6 +1723,159 @@ def two_patches():
...
+               + minus(Dn(u)) * plus(v) )
+    a = BilinearForm((u,v), integral(domain, dot(grad(u),grad(v)))
+                          + integral(I,      expr_I))
+
+#    # TODO BUG
+#    bnd_A = A.get_boundary(axis=0, ext=1)
+#
+#    a = BilinearForm((u,v), integral(domain, dot(grad(u),grad(v)))
+#                          + integral(I,      jump(u) * jump(v))
+#                          + integral(bnd_A,      dx(u)*v))
```
yguclu: Is the bug due to the fact that sympde tries to compute `dx(trace(u))` instead of `trace(dx(u))`?

sympde/topology/domain.py
```diff
@@ -514,6 +515,12 @@ class BoundaryVector(IndexedBase):
 class NormalVector(BoundaryVector):
     pass
 
+class MinusNormalVector(NormalVector):
+    pass
+
+class PlusNormalVector(NormalVector):
```
yguclu: What is the meaning of `MinusNormalVector` and `PlusNormalVector`?

sympde/topology/tests/test_gallery.py
```diff
@@ -44,7 +44,8 @@ def test_square():
     # read it again
     D = Domain.from_file('square.h5')
 
-    assert( D == square )
+    # TODO BUG not working after naming boundaries \Gamma_1 etc
+#    assert( D == square )
```
yguclu: Do we need to update the `export` method of class `Domain`, or the `from_file` constructor?

sympde/expr/tests/test_expr_2d.py
ratnania: it should not compute any of them. 
Right now, we replace all occurrences of a test function `u` by `trace_0(u)` inside `BoundaryIntegral`. This is redundant; it was related to the previous implementation where we didn't have to specify explicitly the boundary as an argument of the integral.

**This will be solved in the issue #41**

sympde/topology/domain.py
ratnania: When having an interface between two subdomains, there will be two normal vectors for each subdomain:
an interface is a tuple (ordered) where the first subdomain (and any expression on it) will have a subscript $-$. The second subdomain will have a $+$
$n_{-}$ is the normal vector to the interface (seen from the first subdomain)
$n_{+}$ is the normal vector to the interface (seen from the second subdomain)

sympde/topology/tests/test_gallery.py
ratnania: Most likely. We normally store the string as `r"\Gamma_1"`. But I don't know what sympy is doing here.

sympde/calculus/core.py
```diff
@@ -1353,6 +1354,383 @@ def eval(cls, *_args):
...
+
+                except:
+                    b = cls(Mul(*vectors), evaluate=False)
+
+            return Mul(a, b)
+
+        return cls(expr, evaluate=False)
+
+#==============================================================================
+class MinusInterfaceOperator(BasicOperator):
```
yguclu: Can you add a short docstring?

sympde/calculus/core.py
```diff
@@ -1353,6 +1354,383 @@ def eval(cls, *_args):
...
+
+            return Dot(Grad(cls(u)), cls(n))
+
+        elif isinstance(expr, NormalVector):
+            return MinusNormalVector('n')
+
+        return cls(expr, evaluate=False)
+
+#==============================================================================
+class PlusInterfaceOperator(BasicOperator):
```
yguclu: Also here, can we add a docstring?

sympde/expr/evaluation.py
```diff
@@ -234,9 +259,228 @@ def _to_matrix_form(expr, M, test_indices, trial_indices):
     if test_indices is None and trial_indices is None:
         return _to_matrix_functional_form(expr, M)
 
+#==============================================================================
+def _split_expr_over_subdomains(expr, interiors, tests=None, trials=None):
```
yguclu: Can we add a docstring with a short explanation of what this function does? Does it return a dictionary of expressions?

sympde/expr/evaluation.py
```diff
@@ -234,9 +259,228 @@ def _to_matrix_form(expr, M, test_indices, trial_indices):
...
+        olds = tests
+        news = d_tests[interior]
+        for old,new in zip(olds, news):
+            d_expr[interior] = d_expr[interior].subs({old: new})
+    # ...
+
+    return d_expr
+
+#==============================================================================
+def _split_expr_over_interface(expr, interface, tests=None, trials=None):
```
yguclu: Here, too, a short docstring would be very helpful.

sympde/topology/tests/test_gallery.py
yguclu: Can you create an issue so that we do not forget about this bug? We will fix it in the future.

sympde/expr/tests/test_expr_2d.py
```diff
@@ -1723,6 +1723,159 @@ def two_patches():
...
+
+    # Nitsch
+    kappa = Constant('kappa')
+    expr_I = ( - jump(u) * jump(Dn(v))
+               + kappa * jump(u) * jump(v)
+               + plus(Dn(u)) * minus(v)
+               + minus(Dn(u)) * plus(v) )
+    a = BilinearForm((u,v), integral(domain, dot(grad(u),grad(v)))
+                          + integral(I,      expr_I))
+
```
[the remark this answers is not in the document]
ratnania: Adding the `assert` here should be done in a separate issue, since we having too many calls to `TerminalExpr` that are not checked using `assert`.
I suggest that we create an issue for that and we go through all our examples.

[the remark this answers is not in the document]
yguclu: Agreed. Please open an issue

sympde/topology/tests/test_topology.py
```diff
@@ -160,8 +160,8 @@ def test_domain_join_line():
                bnd_minus = A.get_boundary(axis=0, ext=1),
                bnd_plus  = B.get_boundary(axis=0, ext=-1))
     ABC = AB.join(C, name = 'ABC',
-               bnd_minus = B.get_boundary('Gamma_2'),
-               bnd_plus  = C.get_boundary('Gamma_1'))
+               bnd_minus = A.get_boundary(axis=0, ext=1),
+               bnd_plus  = B.get_boundary(axis=0, ext=-1))
 
     print(ABC)
```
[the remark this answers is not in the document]
ratnania: a new test was added: `test_boundary_3`

sympde/expr/tests/test_expr_2d.py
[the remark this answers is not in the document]
ratnania: done: issue#43

PR #48, opened 2019-09-17
Analytic mapping

ratnania:


yguclu:
@ratnania Codacy is complaining about the style sheet `notebooks/styles/custom.css`, because it contains HTML instructions while it should not. Of course Codacy is right and we should fix this problem before merging.

Essentially the aforementioned file is not a proper CSS style sheet, but rather an HTML file with inline style information. To solve the problem it should be enough to change the file extension to `.html` and add the document type declaration `<!DOCTYPE html>` at the beginning of the file.

PR #57, opened 2019-11-19
Fix order of (trial, test) functions in Equation class.

yguclu:


yguclu:
@ratnania: I am making some changes that will be needed in Psydac, but by doing so I caused a few unit tests to fail. I am closing this pull request for the time being, I will reopen it later.

yguclu:
I can reopen it now.

PR #58, opened 2019-11-21
Some fixes in Equation, Linear/BilinearForm, Interface and Union classes.

yguclu:


yguclu:
@ratnania: Codacy complains about the syntax
```
values = [*trials, *tests]
```
which seems perfectly valid to me. I think we use this in many places, so I would not be happy to change it. What do you think?

yguclu:
> thanks for these changes/fixes.
> is it possible to add the following verification:
> 
>     * check that the same boundary is not used in the weak formulation and strong condition (essential bc)

Yes, I can definitely add that. But are we totally sure that we will never have such a condition? I am thinking about the biharmonic equation for example...

ratnania:
> @ratnania: Codacy complains about the syntax
> 
> ```
> values = [*trials, *tests]
> ```
> 
> which seems perfectly valid to me. I think we use this in many places, so I would not be happy to change it. What do you think?

I think this has been changed in one of the latest Python versions. I'm not sure if this is true for all py3 versions.

ratnania:
> > thanks for these changes/fixes.
> > is it possible to add the following verification:
> > ```
> > * check that the same boundary is not used in the weak formulation and strong condition (essential bc)
> > ```
> 
> Yes, I can definitely add that. But are we totally sure that we will never have such a condition? I am thinking about the biharmonic equation for example...

Indeed, we need to compute the order of the operator (Laplacian -> 2, etc) to decide either or not we accept multiple bc (weak and strong) on the same boundary.

yguclu:
This kind of verification of the input equation is very interesting, but we need to design it carefully. @saidctb pointed out to me that we should also check that no boundary conditions are applied on a boundary along a periodic direction. I suggest we create a separate issue, or even better a milestone.

ratnania:
> This kind of verification of the input equation is very interesting, but we need to design it carefully. @saidctb pointed out to me that we should also check that no boundary conditions are applied on a boundary along a periodic direction. I suggest we create a separate issue, or even better a milestone.

ok. issue created @issue#59

yguclu:
> > @ratnania: Codacy complains about the syntax
> > ```
> > values = [*trials, *tests]
> > ```
> > 
> > 
> > which seems perfectly valid to me. I think we use this in many places, so I would not be happy to change it. What do you think?
> 
> I think this has been changed in one of the latest Python versions. I'm not sure if this is true for all py3 versions.

Yes, I think you are right. We test our code with Python-3.5, 3.6 and 3.7 and everything works fine. But I remember that some syntax constructs were not accepted on Python-3.4, and this is probably one of them.

From what I could gather Codacy uses Prospector, which in turn uses Pylint. Do you know how to tell Pylint which Python3 version it should accept?

sympde/expr/expr.py
```diff
@@ -502,7 +502,7 @@ def __call__(self, trials, tests, **kwargs):
         if not is_sequence(tests ): tests  = [tests ]
 
         # Concatenate input values into single list
-        values = [*trials, *tests]
```
yguclu: This is not going to work for tuples...

sympde/expr/expr.py
```diff
@@ -502,7 +502,7 @@ def __call__(self, trials, tests, **kwargs):
...
+        values = trials + tests
```
yguclu: @ratnania This is not going to work for tuples!

We should not change perfectly correct code to something not correct, only because Codacy complains about it...

yguclu: In the future Codacy is going to complain about hundreds of situations like this, and we are talking about perfectly legal Python-3.6+ code. So we need to understand how to tell Codacy which Python3 version we are using.

yguclu: I have just asked the Codacy support team about this issue, I will post their answer as soon as I get it.

yguclu: The Codacy guys have replied. I deactivated Prospector and PyLint 1.9.5, and enabled PyLint (Python 3) 2.4.3. Things work now.

PR #61, opened 2020-01-20
[mapping] add SymbolicWeightedVolume class

saidctb:


yguclu:
In the future I think we should move `class SymbolicMappingExpr` and all its subclasses to the Psydac library.

Issue #50, opened 2019-09-26
Remove obsolete classes ScalarField and VectorField


yguclu:
Scalar and vector functions should only be created using `element_of`, which returns objects of type `ScalarTestFunction` and `VectorTestFunction`. Therefore there is no need for the old classes `ScalarField` and `VectorField`, which should be removed from the library.

yguclu:
After merging [Psydac's PR #87](https://github.com/pyccel/psydac/pull/87)  we can proceed with solving this issue.

PR #64, opened 2020-03-02
Devel psydac

saidctb:


sympde/topology/mapping.py
```diff
@@ -776,17 +776,14 @@ def eval(cls, *_args, **kwargs):
 
             return Matrix(lines)
 
-        elif isinstance(expr, (ScalarField, ScalarTestFunction)):
+        elif isinstance(expr, (ScalarField, ScalarTestFunction, VectorField, VectorTestFunction)):
```
yguclu: As a side note, keep in mind that we will remove the types `ScalarField` and `VectorField` in the near future. So please do not use them for you new developments in Psydac. See issue #50.

yguclu:
Codacy complains about an unused import. I can take care of it.

yguclu:
@saidctb Actually Codacy's complaint may have uncovered a conceptual issue.

You have added an import of `VectorFunctionSpace`, but it is never used. Instead, in `_split_test_function(expr)` you create a `ScalarFunctionSpace` for each component of `expr`. Is this because we are implicitly assuming that any `VectorFunctionSpace` is a tensor product of `ScalarFunctionSpaces`?

I think that such an assumption belongs to `psydac`, which uses tensor-product splines, but not to `sympde`, which does not know anything about the discretization.

PR #66, opened 2020-03-24
Use sympy 1.5

saidctb:


yguclu:
Can we delete the `devel-sympy15` branch, or is there something useful that we should keep?

PR #69, opened 2020-06-08
Add InterfaceMapping

saidctb:
InterfaceMapping is used to represent a mapping in the interface, it's needed in the case of two different mappings.
in this PR we have made the following changes : 
- Add InterfaceMapping class
- Use OrderedDict in TerminalExpr instead of Dict to preserve the order of the expressions
- Add AffineMapping
- Add the __add__ magic method in Integral classes

### DIFF
--- a/sympde/topology/basic.py
+++ b/sympde/topology/basic.py
@@ -113,8 +113,8 @@ class Union(BasicDomain):
         for union in unions:
             args += list(union.as_tuple())
 
-        # Sort domains by name
-        args = sorted(args, key=lambda x: x.name)
+        # remove duplicates and Sort domains by name
+        args = sorted(set(args), key=lambda x: x.name)
 
         # a. If the required Union contains no domains, return None;
         # b. If it contains a single domain, return the domain itself;
--- a/sympde/topology/basic.py
+++ b/sympde/topology/basic.py
@@ -115,7 +115,11 @@ class Union(BasicDomain):
             else:
                 new_args.append(i)
 
-        args = new_args
+        args = []
+        for i in new_args:
+            if i not in args:
+                args.append(i)
+
         # a. If the required Union contains no domains, return None;
         # b. If it contains a single domain, return the domain itself;
         # c. If it contains multiple domains, create a Union object.
--- a/sympde/topology/basic.py
+++ b/sympde/topology/basic.py
@@ -108,17 +108,10 @@ class Union(BasicDomain):
             raise ValueError(msg)
 
         # Flatten arguments into a single list of domains
-        new_args = []
-        for i in args:
-            if isinstance(i, Union):
-                new_args += list(i.as_tuple())
-            else:
-                new_args.append(i)
-
-        args = []
-        for i in new_args:
-            if i not in args:
-                args.append(i)
+        unions = [a for a in args if     isinstance(a, Union)]
+        args   = [a for a in args if not isinstance(a, Union)]
+        for union in unions:
+            args += list(union.as_tuple())
 
         # Sort domains by name
         args = sorted(args, key=lambda x: x.name)
--- a/sympde/topology/basic.py
+++ b/sympde/topology/basic.py
@@ -120,6 +120,9 @@ class Union(BasicDomain):
             if i not in args:
                 args.append(i)
 
+        # Sort domains by name
+        args = sorted(args, key=lambda x: x.name)
+
         # a. If the required Union contains no domains, return None;
         # b. If it contains a single domain, return the domain itself;
         # c. If it contains multiple domains, create a Union object.
@@ -161,14 +164,6 @@ class Union(BasicDomain):
         ls = [i for i in self.args]
         return tuple(ls)
 
-    def __eq__(self, a):
-        if isinstance(a, Union):
-            return set(self.as_tuple()) == set(a.as_tuple())
-        else:
-            return False
-
-    def __hash__(self):
-        return hash(self.as_tuple())
 #==============================================================================
 class ProductDomain(BasicDomain):
     def __new__(cls, *args, name=None):
--- a/sympde/topology/basic.py
+++ b/sympde/topology/basic.py
@@ -114,7 +114,7 @@ class Union(BasicDomain):
             args += list(union.as_tuple())
 
         # Sort domains by name
-        args = sorted(args, key=lambda x: x.name)
+        #args = sorted(args, key=lambda x: x.name)
 
         # a. If the required Union contains no domains, return None;
         # b. If it contains a single domain, return the domain itself;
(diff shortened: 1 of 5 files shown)


sympde/topology/mapping.py
```diff
@@ -245,6 +245,25 @@ def _compute_contravariant(self):
     def _compute_hessian(self):
         raise NotImplementedError('TODO')
 
+class InterfaceMapping(Mapping):
```
yguclu, line 248: What do you mean by `InterfaceMapping`? Is it the mapping on an interface, or the interface between two mappings?

sympde/topology/mapping.py
```diff
@@ -257,6 +301,18 @@ class IdentityMapping(Mapping):
...
+class AffineMapping(Mapping):
+    """
+    Represents a 1D/2D/3D Affine Mapping object.
+
+    Examples
+
+    """
+    _expressions = {'x': 'c1 + a11*x1 + a12*x2 + a13*x3',
+                    'y': 'c2 + a21*x1 + a22*x2 + a23*x3',
+                    'z': 'c3 + a31*x1 + a32*x2 + a33*x3'}
```
yguclu, line 314: How would this work in 2D, for instance? By dropping the `z` key and setting `x3 = 0`?

sympde/topology/mapping.py
```diff
@@ -245,6 +260,35 @@ def _compute_contravariant(self):
...
+    def M1(self):
+        return self._args[0]
+
+    @property
+    def M2(self):
+        return self._args[1]
+
+    @property
+    def is_analytical(self):
+        return self.M1.is_analytical
```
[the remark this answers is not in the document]
saidctb: yes I'm not sure if we will allow that or not

sympde/topology/mapping.py
saidctb: yes

sympde/topology/mapping.py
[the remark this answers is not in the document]
yguclu: I think we need to have an error tolerance below which we consider `M1` and `M2` to be compatible.

sympde/topology/mapping.py
yguclu: Does the user need to specify all free parameters in the mapping? Or do you have default values like in Psydac?

sympde/topology/mapping.py
```diff
@@ -245,6 +245,25 @@ def _compute_contravariant(self):
     def _compute_hessian(self):
         raise NotImplementedError('TODO')
 
+class InterfaceMapping(Mapping):
+    def __new__(cls, M1, M2):
+        return Basic.__new__(cls, M1, M2)
```
campospinto, line 275: The names M1 / M2 do not follow the minus/plus terminology used eg in psydac. Is there a reason for that ?

sympde/topology/mapping.py
saidctb: yes he needs to specify them when creating the mapping

sympde/topology/mapping.py
saidctb: I will change then name to follow the same terminology

sympde/topology/mapping.py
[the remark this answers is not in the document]
saidctb: yes for that we need to have access to the logical domain, which we don't for the moment when creating the mappings

saidctb:
> If you look at the analytical mappings in Psydac, there is a clear distinction between the number of logical coordinates `ldim` and the number of physical coordinates `pdim`. We only require that `pdim >= ldim`, and this means that we can handle generic manifolds like a 1D curve or a 2D surface in 3D Euclidean space. Is there a way to do that in SymPDE?
yes I don't see a raison why we can't do it

sympde/topology/mapping.py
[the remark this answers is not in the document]
campospinto: by 'compatible' do you mean 'continuous' ?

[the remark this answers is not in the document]
yguclu: @campospinto We would like them to be continuous, but there is no way to be completely sure unless they are both analytical (in which case we can compare the symbolic expressions). So we will perform some checks and just say that they are "compatible"...

sympde/topology/mapping.py
saidctb: done

sympde/topology/basic.py
```diff
@@ -108,14 +108,14 @@ def __new__(cls, *args):
...
-        args = sorted(args, key=lambda x: x.name)
-
+        new_args = []
+        for i in args:
+            if isinstance(i, Union):
+                new_args += list(i.as_tuple())
+            else:
+                new_args.append(i)
+
+        args = new_args
```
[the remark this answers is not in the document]
saidctb: I changed back the Union to be unordered

sympde/topology/basic.py
```diff
@@ -108,10 +108,17 @@ def __new__(cls, *args):
...
+        for i in args:
+            if isinstance(i, Union):
+                new_args += list(i.as_tuple())
+            else:
+                new_args.append(i)
+
+        args = []
+        for i in new_args:
+            if i not in args:
+                args.append(i)
```
yguclu, line 121: Given that we sort the arguments afterwards, can we leave the original code unchanged?

saidctb: done

sympde/topology/mapping.py
yguclu: Should we change the arguments' names to `minus` and `plus` as well?

saidctb: done

sympde/topology/domain.py
```diff
@@ -57,10 +57,14 @@ def __new__(cls, name, interiors=None, boundaries=None, dim=None,
...
-                for union in unions:
-                    interiors += list(union.as_tuple())
+                new_interiors = []
+                for i in interiors:
+                    if isinstance(i , Union):
+                        new_interiors += list(i.as_tuple())
+                    else:
+                        new_interiors.append(i)
+
+                interiors = new_interiors
```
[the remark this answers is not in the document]
yguclu: I wonder if we should revert to the original code, remove duplicates and sort the interiors. @saidctb What do you think?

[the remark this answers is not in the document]
saidctb: maybe just keep it as it is because we might add OrderedUnion in the futur and in that case it will be important to keep the order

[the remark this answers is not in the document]
yguclu: But shouldn't we remove duplicates? Or maybe raise an error if we find duplicates?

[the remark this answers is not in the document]
yguclu: We should probably improve the constructor of Domain, but we can leave it for later as a separate pull request.

PR #80, opened 2020-09-23
Multi-patch mapping

saidctb:
- We added the Multi-patch mapping class to handle mappings in the multipatch case.
- Fix bug in Grad and LogicalGrad for the VectorTestFunction case.
- Support sympy1.6.


sympde/topology/mapping.py
```diff
@@ -283,8 +283,9 @@ class MultiPatchMapping(Mapping):
 
     def __new__(cls, dic):
         assert isinstance( dic, dict)
-        return Basic.__new__(cls, dict)
+        return Basic.__new__(cls, dic)
```
yguclu, line 286: This was a nasty bug! Well done for finding it!

sympde/topology/mapping.py
```diff
@@ -566,7 +581,7 @@ def eval(cls, F):
         elif rdim == 3:
             expr = LogicalGrad_3d(F)
 
-        return expr
+        return expr.T
```
yguclu, line 584: The fact that we need to transpose the gradients is a matter of definitions.

Would you mind adding our definition of Jacobian matrix (component-wise) to the docstring? I think that's really important!

sympde/topology/derivatives.py
```diff
@@ -951,19 +942,15 @@ def eval(cls, *_args):
...
+        u  = _args[0]
+        du = (dx1(u), dx2(u))
 
+        if isinstance(du[0], (Tuple, Matrix, ImmutableDenseMatrix)):
+            lines = [list(d[:]) for d in du]
         else:
-            v = ImmutableDenseMatrix([[dx1(u)], [dx2(u)]])
+            lines = [[d] for d in du]
 
+        v = ImmutableDenseMatrix(lines)
```
yguclu, line 953: Is this fix needed because the resulting gradient matrix did not have the correct shape?

sympde/topology/derivatives.py
```diff
@@ -951,19 +942,15 @@ def eval(cls, *_args):
...
-            n = len(u)
-            lines = []
-            for i in range(0, n):
-                line = [dx1(u)[0,i], dx2(u)[0,i]]
-                lines.append(line)
-            v = ImmutableDenseMatrix(lines)
+        u  = _args[0]
+        du = (dx1(u), dx2(u))
 
+        if isinstance(du[0], (Tuple, Matrix, ImmutableDenseMatrix)):
```
yguclu, line 948: Do you really need to check `du[0]`, or is it equivalent to check `u`?

sympde/topology/derivatives.py
saidctb: the gradient was not correcty calculated for the the first case but the transpose of it that was calculated

sympde/topology/derivatives.py
saidctb: I need du in this case because it adds the case of u being a `VectorTestFunction`  otherwise it's treated in else and that's where sympy complains

sympde/topology/mapping.py
saidctb: done

sympde/topology/mapping.py
saidctb: this class I created but never used, but we need it, that how I found the error

sympde/topology/derivatives.py
saidctb: the shape was always correct , but sympy1.5 was more strict in the way you create a `Matrix`

yguclu (approved):

sympde/topology/derivatives.py
yguclu: I see, thanks!

PR #82, opened 2020-10-13
Devel mapping

saidctb:
The following changes have been made:
- Add computation of Jacobian matrix and its inverse in creation of mapping;
- Jacobian matrix and its inverse may be explicitly provided by user when creating symbolic mapping;
- Add CallableMapping where we make symbolic mapping callable;
- Allow for pdim (number of physical dimensions) greater than ldim (number of logical dimensions).


yguclu:
@saidctb Is this PR still a draft, or is it ready for review?

sympde/topology/mapping.py
```diff
@@ -86,25 +86,39 @@ class Mapping(BasicMapping):
...
-            if not len(rdim) == 1:
-                raise ValueError('> Expecting a tuple, list, Tuple of length 1')
+        ldim        = kwargs.pop('ldim', cls._ldim)
+        pdim        = kwargs.pop('pdim', cls._pdim)
+        coordinates = kwargs.pop('coordinates', None)
 
-            rdim = rdim[0]
-        elif rdim is None:
-            rdim = cls._rdim
+        dims = [dim, ldim, pdim]
```
yguclu: I am not convinced that `dim` is a useful argument for the constructor, given that we have `ldim` and `pdim` now. Are you worried that writing
```python
domain = Mapping('Omega', ldim=2, pdim=2)
```
is obscure to the user?

saidctb: yes and in all the cases that we use now, ldim = pdim = dim, so this is a way for the user to write it only once because this is the general case , and in the exception he can specify ldim and pdim

yguclu: Maybe we should only use the `dim` argument and allow for two different kinds of usage?
```python
d1 = Mapping('Omega_1', dim=2)       # ldim = 2, pdim = 2
d2 = Mapping('Omega_2', dim=(2, 3))  # ldim = 2, pdim = 3
```

sympde/topology/mapping.py
```diff
@@ -634,7 +659,7 @@ def eval(cls, F, v):
             raise TypeError('> Expecting a tuple, list, Tuple, Matrix')
 
         M   = Jacobian(F).inv().T
```
yguclu, line 661: Clearly the Jacobian matrix is square (and hence possibly invertible) only if `ldim == pdim`. For this transformation to make sense, I think that we should interpret the input mapping as a change of basis, which must be invertible. Therefore we should check that `F.ldim == F.pdim` before computing `M`.

sympde/topology/mapping.py
yguclu: Well, this does not look any clearer than your approach with keyword arguments... Let's leave things as they are.

saidctb: lol I saw this after I changed it, now I deleted it again , yes it's better to keep it as it is

sympde/topology/mapping.py
saidctb: yes I agree

yguclu (approved):

PR #91, opened 2021-01-27
Some bug-fixes for Psydac

yguclu:
In order to avoid problems when assembling vector expressions over the boundary, we do not create `Trace` objects from the expressions anymore. Such a kind of reverse engineering was complex and error prone, and not really needed at the moment.


saidctb (approved):

saidctb:
we have to deactivate some test for the traces.

saidctb (approved):

PR #95, opened 2021-02-08
FEEC multipatch

saidctb:
- Improve the multiple patches objects to handle vector-valued functions
- Add corners computation for 2d geometries
- Split the interfaces first before the evaluation in TerminalExpr
- Change the `TerminalExpr` signature, now it takes `(expr, domain)` as positional arguments
- Change the `LogicalExpr` signature, now it takes `(expr, domain)` as positional arguments
- Update the version

With the new signatures in `TerminalExpr` and  `LogicalExpr`,  the user can define his own coordinates  ,
but this is beyond the scope of this PR


yguclu:
Can you please provide a general description for this PR, with a detailed list of changes?

yguclu (changes requested):
As I understand, this PR contains some bug-fixes that are necessary for running vector multi-patch examples in Psydac. But there is no way to know that things are properly working now by looking at this repository alone.

Can you please provide some symbolic unit tests as guarantee of correctness? You should be able to extract them from Psydac! This will ensure that we are not breaking SymPDE in the future.

sympde/topology/basic.py
```diff
@@ -309,6 +309,44 @@ def logical_domain(self):
...
+
+    def rotate(self, *directions):
+        assert len(directions) == self.dim-1
+
+        if self.dim == 2:
+            if directions[0] == 1:
+                return self
+            elif directions[0] == -1:
+                return self.domain.get_boundary(axis=self.axis, ext=-self.ext)
+
```
campospinto, line 325: this function seems to return either the same boundary edge (in 2D), or the opposite edge. Why is it called `rotate` ?

sympde/topology/basic.py
```diff
@@ -309,6 +309,44 @@ def logical_domain(self):
     def dim(self):
         return self.domain.dim
 
+    @property
+    def neighbours(self):
+        boundaries = [a for a in self.domain.boundary if a.axis !=self.axis]
+        return Union(*boundaries)
+
```
campospinto, line 316: the `neighbours` of a domain boundary face (edge in 2D) could easily refer to the domains that are connected across this face. Here the function seems to be returning all the faces that touch self. Maybe 'neighbour_boundaries' or 'adjacent_boundaries' is a better name ?

sympde/calculus/matrices.py
```diff
@@ -113,7 +119,7 @@ def arg(self):
 
     def _sympystr(self, printer):
         sstr = printer.doprint
-        return '{}.T'.format(sstr(self.arg))
+        return '({}).T'.format(sstr(self.arg))
```
yguclu, line 122: Do we need to add the parentheses if the argument is just a `Matrix`?

saidctb: in the case the argument is just a Matrix, we don't need it, but if it's an expression then yes

yguclu: So maybe we can differentiate between the two cases...?

saidctb: done

sympde/calculus/matrices.py
```diff
@@ -274,13 +289,14 @@ def _sympystr(self, printer):
...
+    return MatSymbolicMul(*x.args)
+def f2(x):
+    return MatSymbolicAdd(*x.args)
 
 Basic._constructor_postprocessor_mapping[MatrixSymbolicExpr] = {
     "Mul": [f1],
     "Add": [f2]
 }
 
+
```
yguclu, line 302: It is a bit cleaner to define some lambda functions instead of `f1` and `f2`, don't you think?
```python
Basic._constructor_postprocessor_mapping[MatrixSymbolicExpr] = {
    "Mul": [lambda x: MatSymbolicMul(*x.args)],
    "Add": [lambda x: MatSymbolicAdd(*x.args)],
}
```

saidctb: done

sympde/expr/evaluation.py
```diff
@@ -414,16 +473,16 @@ def trial(self):
 #==============================================================================
 class TerminalExpr(CalculusFunction):
 
-    def __new__(cls, *args, **options):
+    def __new__(cls, expr, domain, **options):
```
yguclu, line 476: Please add a docstring to this class... it's very important!

sympde/expr/evaluation.py
```diff
@@ -434,218 +493,239 @@ def __getitem__(self, indices, **kw_args):
...
                     # ...
                     try:
-                        domain = _get_domain(a)
-                        if not isinstance(domain, Union):
-                            domain = (domain, )
+                        domains = _get_domain(a)
+                        if not isinstance(domains, Union):
+                            domains = (domains, )
                     except:
                         pass
```
yguclu, line 633: Ehm... why are we ignoring any exception raised by `_get_domain(a)`?

saidctb: we should maybe delete the try-except clause

sympde/topology/mapping.py
```diff
@@ -336,30 +397,29 @@ def _sympystr(self, printer):
...
-        return self._args[2]
+        return self._axis
 
     def _hashable_content(self):
         if self.axis is not None:
-            return (self.mapping, self.axis)
+            return (type(self).__name__, self.mapping, self.axis)
         else:
-            return ('JacobianInverseSymbol', self.mapping)
+            return (type(self).__name__, self.mapping)
```
yguclu, line 422: Is it useful to distinguish between these two cases?

sympde/calculus/core.py
```diff
@@ -666,6 +665,35 @@ def __new__(cls, arg1, arg2, **options):
         return c*obj
 
 #==============================================================================
+class Trace(BasicOperator):
```
yguclu, line 668: I don't think it's a good idea to call this new class `Trace`, given that SymPDE already has a `Trace` class in module `sympde.topology.space`... So please change the name of one of these two classes.

yguclu (changes requested):
These changes look very good to me! I have a few comments

sympde/expr/evaluation.py
saidctb: done

sympde/topology/basic.py
```diff
@@ -332,6 +365,80 @@ def todict(self):
 
         return OrderedDict(sorted(d.items()))
 
+#==============================================================================
+class CornerBoundary(BasicDomain):
+    """
+    Represents an undefined corner over a domain.
+
+    """
+    def __new__(cls, *boundaries):
```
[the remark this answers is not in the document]
saidctb: it only works, in 2d

sympde/topology/mapping.py
saidctb: yes, we can have the two cases in one expression, so the hash must be different

sympde/topology/domain.py
```diff
@@ -422,6 +408,60 @@ def join(self, other, name, bnd_minus=None, bnd_plus=None):
                       mapping=mapping,
                       logical_domain=logical_domain)
 
+    def get_shared_corners(self):
```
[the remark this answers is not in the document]
saidctb: only in 2D

sympde/calculus/core.py
saidctb: yes, I shouldn't have added it, we have `SymbolicTrace` for that

sympde/expr/evaluation.py
```diff
@@ -110,11 +110,11 @@ def _unpack_functions(ls):
...
     return tuple(funcs)
 
+
 #==============================================================================
-def _get_trials_tests_flattened(expr):
+def _get_trials_tests(expr, flatten):
     """
-    Get all scalar trial/test functions in the given expression, with vector
-    functions decomposed into their scalar components.
+    Get all scalar trial/test functions in the given expression.
```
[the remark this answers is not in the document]
saidctb: done

sympde/expr/evaluation.py
```diff
@@ -130,18 +130,22 @@ def _get_trials_tests_flattened(expr):
     tests : tuple
         All scalar test functions in the given expression, with vector
         functions decomposed into their scalar components.
+
+    flatten: Boolean
+        Decompose the vector trial/test functions into their scalar components.
+
```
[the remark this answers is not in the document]
saidctb: done

sympde/expr/evaluation.py
```diff
@@ -180,29 +184,44 @@ def _to_matrix_form(expr, *, trials=None, tests=None):
     tests : iterable
         List of all scalar test functions (after unpacking vector functions).
 
+    domain : Domain
+        domain of expr
```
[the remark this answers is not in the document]
saidctb: done

sympde/calculus/core.py
saidctb: it's fixed

sympde/topology/basic.py
saidctb: In 2d, it's just a 90° rotation, in 3d is more general because we have two directions and not just one

sympde/topology/basic.py
saidctb: fixed

sympde/topology/basic.py
yguclu: I don't see why it's a 90º rotation in the case that `directions[0] == -1`. It looks like a 180º rotation to me...

sympde/expr/evaluation.py
yguclu: Yes please!

sympde/topology/basic.py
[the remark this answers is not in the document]
yguclu: I see. Please expand the docstring to explain it!

sympde/topology/mapping.py
yguclu: But can't you just use the following in both cases?
```python
return (type(self).__name__, self.mapping, self.axis)
```
The hash will be different if only one of the two cases has `self.axis == None`.

sympde/calculus/core.py
yguclu: Should you then update the PR description and remove the line "Add matrix trace operator"?

saidctb: yes, I will do it

sympde/expr/evaluation.py
```diff
@@ -472,10 +475,26 @@ def trial(self):
...
+          The mathematical expression
+
+        domain: Domain
+          The domain in which the expression is defined
+
+        Returns
+        -------
+          r: sympy.Expr or tuple of sympy.Expr
+            The atomized expression
+        """
```
yguclu, line 497: The constructor's docstrings should be part of the class' docstring, otherwise there is no way to access it from the interpreter...

sympde/topology/mapping.py
saidctb: it will raise an error, if you include None in the tuple

sympde/topology/basic.py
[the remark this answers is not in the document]
saidctb: this can be generalized to 3d, it's just not implemented , should we still do that ?

yguclu (changes requested):

sympde/expr/evaluation.py
saidctb: done

sympde/topology/mapping.py
yguclu: I thought `None` was hashable...

sympde/expr/evaluation.py
saidctb: done

sympde/topology/basic.py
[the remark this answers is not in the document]
yguclu: It depends. If you generalized it to 3D, what would it represent? 0D corners or 1D edges?

sympde/expr/evaluation.py
```diff
@@ -476,25 +476,24 @@ def trial(self):
 #==============================================================================
 class TerminalExpr(CalculusFunction):
     """
-    This class takes a Sympy expression and transforms its vectorized operators into atomic ones.
+     This class takes a Sympy expression and transforms its vectorized operators into atomic ones.
+     in a specified domain.
```
yguclu, line 480: 1. vectorized -> vector
2. Sympy -> SymPDE
3. Maybe you wanted to have a single sentence here?

yguclu (changes requested):

sympde/topology/basic.py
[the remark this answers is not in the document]
saidctb: 1D edges

sympde/topology/mapping.py
saidctb: yes but sympy doesn't like it for some reason

sympde/expr/evaluation.py
saidctb: ah yes, you are right

sympde/topology/basic.py
saidctb: yes my bad 180°

sympde/expr/evaluation.py
saidctb: fixed

sympde/topology/mapping.py
yguclu: Never mind then!

sympde/topology/basic.py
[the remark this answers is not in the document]
yguclu: If we want to have it general, we need to decide what it represents in 1D, 2D and 3D. I think that we should have a look at this whole file together with @campospinto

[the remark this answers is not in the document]
saidctb: done

sympde/topology/domain.py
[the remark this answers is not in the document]
saidctb: I have added the docstrings

yguclu (approved):

PR #101, opened 2022-01-18
Replace OrderedDict with dict. Remove yamlloader dependency

anlavandier:
As of Python 3.7, `dict` objects are now insertion-ordered. For this reason, we can replace all `OrderedDict` objects with standard `dict`s. Consequently, support for Python 3.6 and lower is stopped and the `yamlloader` package is not needed anymore.


.github/workflows/continuous-integration.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.5, 3.6, 3.7, 3.8]
+        python-version: [3.7, 3.8]
```
yguclu, line 17: Please add versions 3.9 and 3.10, as is done in Pyccel

sympde/topology/domain.py
```diff
@@ -260,15 +258,15 @@ def todict(self):
              'boundary':     boundary,
              'connectivity': connectivity}
 
-        return OrderedDict(sorted(d.items()))
+        return {k: v for k,v in sorted(d.items())}
```
yguclu: This also works:
```suggestion
        return dict(sorted(d.items()))
```

yguclu: You were faster than me :smile:

sympde/topology/basic.py
```diff
@@ -105,7 +105,7 @@ def _sympystr(self, printer):
     def todict(self):
         name   = str(self.name)
         d = {'name': name}
-        return OrderedDict(sorted(d.items()))
+        return dict(sorted(d.items()))
```
yguclu, line 108: Not strictly related to this PR, but I take the opportunity to point out that this could be written in one line:
```python
    def todict(self):
        return {'name': str(self.name)}
```

sympde/topology/basic.py
```diff
@@ -363,7 +363,7 @@ def todict(self):
              'axis':  axis,
              'ext':   ext}
 
-        return OrderedDict(sorted(d.items()))
+        return dict(sorted(d.items()))
```
yguclu, line 366: This method can also be simplified quite a bit:
```python
    def todict(self):
        return {'axis'  : str(self.axis),
                'domain': str(self.domain.name),
                'ext'   : str(self.ext),
                'name'  : str(self.name)}

File: sympde/topology/basic.py [hunk 5]
```diff
@@ -535,18 +535,18 @@ def patches(self):
...
+        data = dict(sorted(self._data.items()))
         for _,v in data.items():
             ls.append(v)
         return Union(*ls)
 
     def todict(self):
         # ... create the connectivity
         connectivity = {}
-        data = OrderedDict(sorted(self._data.items()))
+        data = {k: v for k,v in sorted(self._data.items())}
```
yguclu, line 546: As you have done in other places, you can just write
```suggestion
        data = dict(sorted(self._data.items()))
```

.github/workflows/continuous-integration.yml
yguclu: Thanks

sympde/topology/basic.py
anlavandier: I must have missed it because my last commit was all about putting `dict(...)` instead of `{k: v ....}`

yguclu (approved):

PR #111, opened 2022-06-30
Improve calculus operators

saidctb:
- Generalize the div operator to take matrices instead of vectors only.
- Return a column vector instead of a row vector when calling TerminalExpr of a Normal Vector.


yguclu:
@saidctb Can you please add a description to this PR?

saidctb:
> @saidctb Can you please add a description to this PR?

done

saidctb:
> @saidctb Do you need these changes to be available on PyPI as a new SymPDE version?

Yes, we need it in PyPI for the Psydac [PR](https://github.com/pyccel/psydac/pull/213)

yguclu (approved):

PR #118, opened 2022-09-23
Fix bug in domain.py when boundary is empty

vcarlier:
Minor change to prevent bug in case the domain has no boundary when converting it to tuple


yguclu (commented):
@vcarlier Thanks for the bug fix!

In order to prove that the bug is indeed fixed, and to make sure that future changes will not break the new code, you should add a unit test. This can be done in [this file](https://github.com/pyccel/sympde/blob/master/sympde/topology/tests/test_topology.py).

yguclu (approved):

PR #122, opened 2022-12-07
GLT symbol of bilinear forms containing boundary integrals

e-moral-sanchez:
The GLT symbol was not computed correctly when a bilinear form contained a boundary integral. The desired behaviour is:

- `gelatize` of a bilinear form involving only boundary integrals should return zero.
- in the case of having a bilinear form involving the sum of boundary integrals and domain integrals, `gelatize` should only be applied to the sum of domain integrals.


e-moral-sanchez:
Hi @yguclu , after merging the master branch of Sympde into `glt_boundary_fix`, the function `gelatize` does no longer work. This is because `gelatize` calls `sympde.expr.TensorExpr` without passing the domain. Thus before this PR I need to do a PR in Gelato.

The bug was that the GLT symbol of a bilinear form consisting of a boundary integral should be zero. I believe that this test should be in Gelato, although it requires a change in Sympde. What do you think?

yguclu:
If you think that it is easy to fix this, please make a small PR in GeLaTo.

e-moral-sanchez:
The fix was done 1 year ago for the new version of GeLaTo (version `0.12.0`). However Psydac requires Gelato version `0.11`, which does have this fix.

yguclu:
> The fix was done 1 year ago for the new version of GeLaTo (version `0.12.0`). However Psydac requires Gelato version `0.11`, which does have this fix.

I am not sure that I follow here. Did you mean "which does **not** have this fix?" Will it be sufficient to change the GeLaTo version used by Psydac? Please explain the (ordered) steps which should be taken in order to close this issue.

In any case this PR was created in December 2022 **after** the deployment of GeLaTo version 0.12.0 (June 2022)...

e-moral-sanchez:
Yes, sorry, I meant "not".  The problem is:
- With Gelato version `0.11` the function `gelatize` does not work (it does not include this fix)
- With Gelato version `0.12.0` the function `gelatize` does work (it includes this fix)
- Psydac requires Gelato version `0.11`.

These are the steps I would follow:
1. Update Psydac's requirement to allow Gelato version `0.12.0` if possible
2. Test the `gelatize` function with the changes in this PR and make sure that the GLT symbol of a bilinear form involving a boundary integral is zero. I may need to add some line either in Sympde or Gelato.
3. Add a test in Gelato to make sure that it works correctly.

yguclu:
> In order to have the new unit tests in Gelato passing, we need to merge these changes first.

If GeLaTo requires specific changes in SymPDE, we should update the version of SymPDE so I can upload it to PyPI. Then GeLaTo will have to use the new version.

sympde/expr/evaluation.py
```diff
@@ -1140,7 +1140,7 @@ def eval(cls, *_args, **kwargs):
 
             return ImmutableDenseMatrix(lines)
 
-        elif isinstance(expr, DomainExpression):
+        elif isinstance(expr, DomainExpression) or isinstance(expr, BoundaryExpression):
```
yguclu, line 1143: The intrinsic function `isinstance` accepts a tuple of types as second argument:
```suggestion
        elif isinstance(expr, (DomainExpression, BoundaryExpression)):
```

yguclu (changes requested):
Please update the library version to 0.17.1

sympde/expr/evaluation.py
```diff
@@ -1160,7 +1160,7 @@ def eval(cls, *_args, **kwargs):
             # ...
 
             # ...
-            if domain is not None and domain.mapping is not None:
+            if domain is not None and hasattr(domain, 'mapping') and domain.mapping is not None:
```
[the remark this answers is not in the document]
e-moral-sanchez: 28a2cc124c130a41930b5c786ab9d5ef3b610aed

sympde/expr/evaluation.py
e-moral-sanchez: 28a2cc124c130a41930b5c786ab9d5ef3b610aed

yguclu (approved):
Can you please add a description to this PR?

The title might also need to be updated.

PR #136, opened 2023-07-07
Fix bounds bug in `plot_3d_single_patch`

anlavandier:
This PR applies a one character fix in the function `plot_3d_single_patch` which would previously calculate the surface $\lbrace (x, y_{\text{max}}, z) | (x, z) \in [x_{\text{min}}, x_{\text{max}}] \times [z_{\text{min}}, z_{\text{max}}] \rbrace$ using $x_{\text{max}}$ instead of $y_{\text{max}}$.


anlavandier:
Hi, 
I was looking through some of psydac and SymPDE's code and found this little mistake.

yguclu (approved):
Looks good. Thanks!

PR #139, opened 2023-07-27
Add SemiNorm

saidctb:
Rename `Norm` to `SemiNorm` and add a new `Norm` class.

This is important because, for example, the H1-norm is equal to the L2-norm plus the H1-seminorm.


sympde/expr/expr.py
yguclu, line 1: There is quite a bit of code duplication between the `Norm` and the `SemiNorm`. Is there no way to avoid that?

sympde/expr/expr.py
```diff
@@ -599,6 +600,89 @@ def kind(self):
...
+#        # ...
+#        tests = expr.atoms((ScalarFunction, VectorFunction))
+#        if tests:
+#            msg = '> Expecting an Expression without test functions'
+#            raise UnconsistentArgumentsError(msg)
+#
+#        if not isinstance(expr, (Expr, Matrix, ImmutableDenseMatrix)):
+#            msg = '> Expecting Expr, Matrix, ImmutableDenseMatrix'
+#            raise UnconsistentArgumentsError(msg)
+#        # ...
```
yguclu, line 616: Could this block of code be useful at some point? If not, please remove it.

saidctb: it is not useful

sympde/expr/expr.py
saidctb: We probably can create a superclass and pass a flag to it, but it will have ifs there, so I'm not sure if we can make it better

sympde/expr/expr.py
saidctb: fixed in [fd83e14](https://github.com/pyccel/sympde/pull/139/commits/fd83e1436a986e1abb3a39e87cb39707233099fa)

sympde/expr/expr.py
yguclu: Actually I was thinking that every `Norm` is also a `SemiNorm`, which in turn is a `Functional`. So maybe there is no need for a superclass?

saidctb: so how would we do it?

yguclu: Never mind, I think that we have to clean up various things before worrying about this one. For starters, the `Functional` class currently applies an integral to its argument, which is inconsistent with `LinearForm` and `BilinearForm`.

yguclu (approved):

Issue #145, opened 2024-02-18
algebraic operators do not handle sympy/Matrix


ratnania:
The algebraic operators `dot`, `cross`, `inner` and `outer` do not handle correctly `Matrix` objects from sympy.

We consider the following 

```python
from sympde.topology import VectorFunctionSpace, Cube, element_of
from sympde.calculus import grad, dot, inner, outer, cross, div, Transpose

from sympy import Matrix

domain = Cube()

I = Matrix([[1, 2, 3], [0, 1, 0], [0, 0, 1]])

V = VectorFunctionSpace('V', domain)

u,v = [element_of(V, name=i) for i in ['u', 'v']]
```

Then all the following operations lead to the same error

```python
>>> A = Matrix([[1, 2, 3], [2, 1, 0], [3, 0, 1]])

>>> dot(A*u, v)
...
AttributeError: 'MutableDenseMatrix' object has no attribute 'is_commutative'
>>> cross(A*u, v)
...
AttributeError: 'MutableDenseMatrix' object has no attribute 'is_commutative'
>>> outer(A*u, v)
...
AttributeError: 'MutableDenseMatrix' object has no attribute 'is_commutative'

>>> D = Matrix([[div(u), 0, 0], [0, div(u), 0], [0, 0, div(u)]])

>>> inner(D,D)
...
AttributeError: 'MutableDenseMatrix' object has no attribute 'is_commutative'
```

The problem comes from the same piece of code when treating the `Mul` objects;

```python
        if isinstance(arg1, Mul):
            a = arg1.args
        else:
            a = [arg1]

        if isinstance(arg2, Mul):
            b = arg2.args
        else:
            b = [arg2]

        args_1 = [i for i in a if not i.is_commutative]
        c1     = [i for i in a if not i in args_1]
        args_2 = [i for i in b if not i.is_commutative]
        c2     = [i for i in b if not i in args_2]

        a = reduce(mul, args_1)
        b = reduce(mul, args_2)
        c = Mul(*c1)*Mul(*c2)
```

e-moral-sanchez:
We want to be able to assemble the bilinear form $\int_{\Omega} (\boldsymbol{b} \times \boldsymbol{u}) \cdot \boldsymbol{v} d \boldsymbol{x}$ using the [cross product matrix](https://github.com/pyccel/sympde/issues/146#issue-2144504549).
[...]

PR #147, opened 2024-02-23
Devel matrices

ratnania:
This PR solves #145 issue.

In this PR, we introduce the objects `Vector` and `Matrix`. They are meant to be immutable and only as an API for the user.
All examples using `Tuple` as an element of user expressions have been updated, and we should from now on, remove the use of `Tuple` as it not an algebraic object.

Internally, we convert these new objects to `sympy`, either through the objects method `to_sympy()` or when evaluating symbolic expressions, for instance inside `TerminalExpr` or mapping evaluation operators.


sympde/topology/tests/test_derivatives.py
```diff
@@ -54,14 +55,14 @@ def test_partial_derivatives_1():
...
-                        alpha*dx(v) + beta*dy(v),
-                        alpha*dx(w) + beta*dy(w)]])
+    assert(dx(uvw) == Matrix([[dx(u)], [dx(v)], [dx(w)]]))
+    assert(dx(uvw) + dy(uvw) == Matrix([[dx(u) + dy(u)],
+                                        [dx(v) + dy(v)],
+                                        [dx(w) + dy(w)]]))
+
+    expected = Matrix([[alpha*dx(u) + beta*dy(u)],
+                       [alpha*dx(v) + beta*dy(v)],
+                       [alpha*dx(w) + beta*dy(w)]])
```
yguclu, line 65: The original test was expecting a row vector, while now a column vector is created?

sympde/expr/tests/test_expr.py
```diff
@@ -1891,6 +1895,200 @@ def test_interface_integral_4():
...
+    a1    = Constant('a1',    is_real=True)
+    a2    = Constant('a2',    is_real=True)
+    b1    = Constant('b1',    is_real=True)
+    b2    = Constant('b2',    is_real=True)
+
+    V = ScalarFunctionSpace('V', domain)
+    W = VectorFunctionSpace('W', domain)
+
+    u,v   = [element_of(V, name=i) for i in ['u', 'v']]
+    F,G   = [element_of(W, name=i) for i in ['F', 'G']]
```
yguclu, line 1918: Shorter and more elegant:
```suggestion
    u, v   = elements_of(V, names='u, v')
    F, G   = elements_of(W, names='F, G')
```

yguclu (changes requested):
This is a great improvement!!

I have a few comments and questions.

sympde/topology/tests/test_derivatives.py
ratnania: This was a bug in sympde. 
Let's take the following `sympy` code
```python
from sympy import Matrix

A = Matrix([[1,2], [3,4]])
x = Matrix([[-1,-1]])
```
this is in fact, how the `Tuple` was converted before.
Then we get the following error,
```python
/usr/lib/python3/dist-packages/sympy/matrices/common.py in multiply(self, other, dotprodsimp)
   2783              getattr(other, 'is_MatrixLike', True))):
   2784             if self.shape[1] != other.shape[0]:
-> 2785                 raise ShapeError("Matrix size mismatch: %s * %s." % (
   2786                     self.shape, other.shape))
   2787 

ShapeError: Matrix size mismatch: (2, 2) * (1, 2).
```
There is a mismatch in the Matrix size.
The new implementation, fixes this problem, since we should implement a `Tuple` or a `Vector` as
```python
from sympy import Matrix

A = Matrix([[1,2], [3,4]])
x = Matrix([[-1],[-1]])
```

sympde/topology/derivatives.py
```diff
@@ -82,6 +84,10 @@ def eval(cls, expr):
             args = [cls(i, evaluate=True) for i in expr]
             args = Tuple(*args)
             return Matrix([args])
+        elif isinstance(expr, Vector):
+            args = [[cls(_, evaluate=True)] for _ in expr.args[0]]
+            return ImmutableDenseMatrix(args)
```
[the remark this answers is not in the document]
ratnania: fixed

sympde/core/matrices.py
```diff
@@ -289,6 +293,124 @@ def _sympystr(self, printer):
...
+    @property
+    def name(self):
+        return self._name
+
+    def __getitem__(self, key):
+        try:
+            i,j = key
+            return self.args[0][i][j]
+        except:
+            return MatrixElement(self, key)
```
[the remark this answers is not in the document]
ratnania: You're right, we don't need it, since our `Matrix` & `Vector` are meant to be immutable.
Initially, I didn't plan to have them this way, but after playing with it and using it, I think it is better to stick to immutable objects. Then the `try/except` must be removed.

sympde/core/matrices.py
```diff
@@ -289,6 +293,124 @@ def _sympystr(self, printer):
...
+        return '{}'.format(sstr(self.name))
+
+    def __hash__(self):
+        return hash((self.name, self.args))
+
+    def __eq__(self, a):
+        if isinstance(a, Matrix):
+            eq = self.name == a.name
+            return eq
+        return False
```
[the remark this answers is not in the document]
ratnania: The test on the Vector/Matrix elements has been added

sympde/core/matrices.py
```diff
@@ -289,6 +293,124 @@ def _sympystr(self, printer):
...
+            a = Tuple(*_)
+            newargs.append(a)
+        _args = Tuple(*newargs)
+
+        args = (_args, args[1:])
+
+        obj = Expr.__new__(cls, *args)
+        obj._name = name
+
+        return obj
```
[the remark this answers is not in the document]
ratnania: Let's go for this one.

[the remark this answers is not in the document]
ratnania: same thing has been done for the `Vector` class.

ratnania:
There's still a problem with the `cross` operator. I'm working on it.

ratnania:
The `cross` problem has been fixed;
```python
        # after simplifications, the expression can be reduced to 0 in some cases
        # here's an example;
        # if b is a Vector and u and v are scalar functions,
        # we should have the following
        # cross(b*u,b*v) = u*v*cross(b,b) = 0
        if a == b:
            return S.Zero
```
ready to merge.

sympde/core/matrices.py
```diff
@@ -289,6 +295,106 @@ def _sympystr(self, printer):
         sstr = printer.doprint
         return '{}[{}]'.format(sstr(self.args[0]),sstr(self.args[1]))
 
+class Matrix(MatrixSymbolicExpr):
+    def __new__(cls, mat, *, name):
```
yguclu, line 299: Looking at the code one more time, I find myself unsure about the necessity of giving a "name" to the class instances. This is not a purely symbolic class, but rather a container for symbolic objects. So why do we care about giving the objects a string name?

sympde/expr/tests/test_expr.py
```diff
@@ -1878,19 +1882,232 @@ def test_interface_integral_4():
...
+    #A = SympdeMatrix([[cos(theta), 0], [0, sin(theta)]], name='A')
+    A = SympdeMatrix([[1, 2, 3], [3, 4, 5], [5, 6, 7]], name='A')
+    b = Vector([1, 2, 3], name='b')
+
+    # ...
+    b1 = b
+    assert b1 == b
+
+    b1 = Vector([1, 2, 3], name='b1')
+    assert not b1 == b
```
yguclu, line 1927: See my previous comment. It is quite strange that by comparing two identical vectors the equality is not satisfied, because they have different names. Then what is the meaning of equality for this class?

sympde/core/matrices.py
ratnania: In fact, both `Vector` and `Matrix` should be used in two different ways;
- as a container; like the `Tuple`, but allow for algebraic operations, which is not the case of `Tuple`. In this case, we don't need them to have the `name` attribute.
- as a variable; like the `Constant`, in which case, we'll be able to pass it as an argument when doing the assembly. This will make our code much more readable.
This means that the actual implementation needs improvements (add `is_real` or `is_complex` etc then update psydac).
So I suggest that we keep it this way, then we'll open an issue to make them work as variables.

sympde/expr/tests/test_expr.py
ratnania: see my remark on the `Matrix` case.

PR #153, opened 2024-04-18
Allow turning off linearity checks in bilinear and linear forms

e-moral-sanchez:
Add the boolean flag `check_linearity` to the constructors of `BilinearForm` and `LinearForm`.  The default value is `True`. If set to `False`, the linearity checks are turned off completely. This is different than the existing flag `ignore_linearity_errors`, which controls whether we get a warning or an error when the linearity check fails. 

The new flag might be useful when the expression in the bilinear or linear form is complicated, because then the linearity check calls the SymPy `Expr` method `.expand(deep=True)`, which is recursive and can take a very long time.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name            = "sympde"
-version         = "0.18.2"
+version         = "0.18.3"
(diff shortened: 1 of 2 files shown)


yguclu (approved):
Looks good to me!

Issue #140, opened 2023-09-20
Add docstring to Domain.join()


yguclu:
The method `join` of class `Domain`, albeit very powerful, is not documented. At the very least an extended docstring should be provided.

PR #155, opened 2024-06-03
improving Domain.join

campospinto:
I started a docstring for Domain.join(), and also added the possibility of passing the patch objects (not just their local list indices) in the connectivity list.

This should fix issue [#140](https://github.com/pyccel/sympde/issues/140)

- [x] some pieces of the docstring need to be further clarified

### DIFF
--- a/sympde/topology/domain.py
+++ b/sympde/topology/domain.py
@@ -463,7 +463,7 @@ class Domain(BasicDomain):
         # the multi-patch domain
         Omega = Domain.join(patches=patches, connectivity=connectivity, name='Omega')
         
-        """        
+        """
 
         assert isinstance(patches, (tuple, list))
         assert isinstance(connectivity, (tuple, list))
--- a/sympde/topology/domain.py
+++ b/sympde/topology/domain.py
@@ -411,7 +411,8 @@ class Domain(BasicDomain):
             and 
             - In 2D, ornt is an integer that can take the value of 1 or -1
             - In 3D, ornt is a tuple of 3 integers that can take the value of 1 or -1
-            TODO: specify how these orientations are defined -- also not clear in the example below
+            (see below for more details)
+            
         
         name : str
             name of the domain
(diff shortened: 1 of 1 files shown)


sympde/topology/domain.py
```diff
@@ -403,13 +472,21 @@ def join(cls, patches, connectivity, name):
         assert all(p.dim==patches[0].dim for p in patches)
         dim = int(patches[0].dim)
 
+        patch_given_by_indices = (len(connectivity) > 0 and isinstance(connectivity[0][0][0], int))
```
saidctb, line 475: In which case will the patches be given by the connectivity and not be passed as a list?

campospinto: patches are always passed as a list. we are talking about how the patches are passed in the connectivity argument here: they can be passed
 - either by their indices in the list (as it was done before)
 - or as objects (which allows for a clearer syntax when the function is called: this is the example in the docstring)
I can specify this in the docstring

campospinto: An important missing information, though, is how the orientation is defined. I could not find this information.

saidctb: For the orientation you can check the reference specified in the docstring
T. Dokken, E. Quak, V. Skytt. Requirements from Isogeometric Analysis for changes in product design ontologies, 2010.
The orientations are specified in the same manner as in GeoPDES

campospinto: I just checked this reference but could not find a simple definition of the orientation triple in 3D. We should recall this definition and provide a simple example

saidctb: [Here](https://github.com/rafavzqz/geopdes/blob/master/geopdes/doc/geo_specs_mp_v21.txt#L193-L237) you can find a better explanation with examples

campospinto: Thanks, I'm adding this link to the docstring.

campospinto: @saidctb do you have further questions ?

saidctb: No, it looks good to me

saidctb (approved):

PR #156, opened 2024-06-14
Domain coordinates should be real

e-moral-sanchez:
At the moment the domain coordinates are Sympy symbols but we don't specify them to be real.

Consequently when we take the real part of a complex expression of the coordinates, the resulting symbolic expression includes the imaginary part of the coordinate and eventually Pyccel throws an error.

Besides, setting the symbols to real simplifies considerably the expressions.


campospinto (approved):
This seems indeed a good change. I don't see any practical issue.

PR #158, opened 2024-07-15
adding test with numpy constants

campospinto:
On my machine these lines trigger an error when run with python3.12

```
    import numpy as np
    from sympde.topology import AffineMapping

    dim   = 2
    alpha = np.pi/2
    c1    = 0.2
    c2    = 1.5

    F =  AffineMapping(
        name='F', dim=2, c1=c1, c2=c2,
        a11=np.cos(alpha), a12=-np.sin(alpha),
        a21=np.sin(alpha), a22=np.cos(alpha),
    )
```

### DIFF
--- a/sympde/topology/mapping.py
+++ b/sympde/topology/mapping.py
@@ -223,7 +223,7 @@ class Mapping(BasicMapping):
             constants_values = {a.name:Constant(a.name) for a in constants}
             # subs constants as Constant objects instead of Symbol
             constants_values.update( kwargs )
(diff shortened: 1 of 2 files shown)


campospinto:
One solution seems to convert constants to floats in L 226 of `sympde/topology/mapping.py`:
that is, replace 
```
            d = {a:constants_values[a.name] for a in constants}
```
by 
```
            d = {a:float(constants_values[a.name]) for a in constants}
```

saidctb (approved):

PR #161, opened 2024-07-16
Add tests for Navier-Stokes terminal expressions 

campospinto:
A couple of tests are added for terminal expressions used in the Navier-Stokes equations. This is useful to verify exact solutions, e.g. in the Psydac tests.

In Psydac this code was tested in the "testing" PR pyccel/psydac#464.


sympde/expr/tests/test_expr.py
```diff
@@ -1890,6 +1890,59 @@ def test_interface_integral_4():
...
+
+    fx = -mu*(ux.diff(x, 2) + ux.diff(y, 2)) + ux*ux.diff(x) + uy*ux.diff(y) + pe.diff(x)
+    fy = -mu*(uy.diff(x, 2) + uy.diff(y, 2)) + ux*uy.diff(x) + uy*uy.diff(y) + pe.diff(y)
+
+    # [MCP 18.07.2024] for now, this test fails here because f is essentially 0: this should be fixed
+    print(f'a = {a}')
+    print(f'b = {b}')
+    print(f'f = {f}')
+    assert (f[0]-fx).simplify() == 0
+    assert (f[1]-fy).simplify() == 0
```
yguclu, line 1925: What does the comment line mean? Which test fails? Both assert statements pass

sympde/expr/tests/test_expr.py
```diff
@@ -1890,6 +1890,59 @@ def test_interface_integral_4():
...
+
+    mu = 1
+    ux = cos(y*pi)
+    uy = x*(x-1)
+    ue = Matrix([[ux], [uy]])
+    pe = sin(pi*y)
+    # ...
+
+    # Verify that div(u) = 0
+    assert (ux.diff(x) + uy.diff(y)).simplify() == 0
```
yguclu, line 1908: Could we use `div(ue)` and `TerminalExpr` here?
```python
d = TerminalExpr(div(ue), domain)
assert d.simplify() == 0

File: sympde/expr/tests/test_expr.py (hunk 1 above)
Comment (campospinto) [reply sympde/expr/tests/test_expr.py:1925]: Indeed, these statements pass now, maybe the problem was fixed by merging master into this branch.

... but something is still strange: if I run this test directly (with `$ python sympde/expr/tests/test_expr.py`), then the  assert fails:

```
a = Matrix([[0], [0]])
b = Matrix([[0, 0], [0, 0]])
f = Matrix([[0], [0]])
Traceback (most recent call last):
  File "/Users/campos/Work/codes/pyccel/sympde/sympde/expr/tests/test_expr.py", line 1962, in <module>
    test_terminal_expressions_for_navier_stokes()
  File "/Users/campos/Work/codes/pyccel/sympde/sympde/expr/tests/test_expr.py", line 1924, in test_terminal_expressions_for_navier_stokes
    assert (f[0]-fx).simplify() == 0
AssertionError

```

sympde/expr/tests/test_expr.py
campospinto: I think so: I added this in the test which seems to be passing

sympde/expr/tests/test_expr.py
campospinto: Ok, it seems that this failed assert with direct python call (no pytest) was due to an older version of sympde -- for some reason pytest was importing the modules from a different place...

So the current code should be fine 👍

yguclu:
Waiting for the tests to pass in pyccel/psydac#464...

yguclu:
All tests in pyccel/psydac#464 are passing now.

PR #163, opened 2024-07-24
reverting real coordinates

campospinto:
The PR #156 defined domain coordinates with real=True argument. This is probably justified but requires additional testing as it breaks some calls to TerminalExpr. Here this change is reverted, and a dedicated PR will be reopened later.

### DIFF
--- a/sympde/topology/domain.py
+++ b/sympde/topology/domain.py
@@ -839,7 +839,7 @@ class NCube(Domain):
             raise ValueError("Min coordinates must be smaller than max")
 
         coord_names = 'x1:{}'.format(dim + 1)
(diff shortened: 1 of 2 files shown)


e-moral-sanchez (approved):
Ok. I will take a look at this later on. Thanks!

yguclu:
@campospinto @e-moral-sanchez What is the current status on this matter? Is there an open issue?

campospinto:
> @campospinto @e-moral-sanchez What is the current status on this matter? Is there an open issue?

I don't think so, but it would be good indeed to have one that describes precisely the problem. 

Note that the previous PR #156 was breaking some calls to TerminalExpr of derivatives (returning zeros for non-constant functions), probably because the coordinate symbols were not properly identified. Some of the corresponding code is tested  in the (still open) PR #161

PR #172, opened 2025-02-27
Improve installation instructions in README

yguclu:
* Recommend virtual environment with instructions
* Explain usage of `pip` and PyPI
* Provide `git` commands
* Do not suggest `--user` flag to `pip` command


campospinto (approved):
This is very useful, thanks!

PR #173, opened 2025-03-04
Unique API for single/multi-patch `Domain` objects

### LINKED ISSUE #171: Unify `Domain` behavior for single and multi-patch cases
In Psydac it is difficult to use a `Domain` object in the single and multi-patch cases without handling lots of corner cases with `if` statements. One reason is that `domain.interior.args` returns objects of different types:

   - for a multi-patch domain: the subdomains as a tuple of logical domains 
   - for a single patch domain: the name of the unique subdomain (the domain itself), as a tuple of strings, 
      
This is probably because `domain.interior` is a Union of subdomains.

To fix this without changing how Unions are handled in SymPDE, we could add a function/attribute `domain.subdomains` which always returns the subdomains as a tuple of logical domains.

campospinto:
In order to improve the compatibility between single and multi-patch domains, we add:

- `subdomains` and `mappings` attributes to the `Domain` class, which always return a tuple (fixes #171);
- the possibility of calling the `join` function (which usually creates multipatch domains) with a single patch.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name            = "sympde"
-version         = "0.19.0"
+version         = "0.19.1"
 description     = "Symbolic calculus for partial differential equations (and variational forms)"
 readme          = "README.rst"
 requires-python = ">= 3.8, < 3.13"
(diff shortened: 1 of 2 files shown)


campospinto:
see associated testing PR in psydac: https://github.com/pyccel/psydac/pull/466

yguclu (approved):

yguclu:
@campospinto @e-moral-sanchez With this PR we will increase SymPDE's version to 0.19.1.

PR #177, opened 2025-03-25
Stop support of Python 3.8, add support of Python 3.13

### LINKED ISSUE #176: Add support for Python 3.13
Since Python 3.13 is quickly becoming the default in many `:latest` images, how difficult is it to lift the requirement `Python<3.13`?

yguclu:
Fixes #176. On 7 October 2024:
* Python 3.8 reached end of life;
* Python 3.13 was first released.

See https://devguide.python.org/versions/ for reference.

Hence we drop here support for Python 3.8, and allow for installation with any Python version >= 3.9. Unit tests are now run with all Python versions between 3.9 and 3.13, on Ubuntu and macOS runners.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -9,13 +9,13 @@ on:
     branches: [ master ]
 
 jobs:
-  Linux:
-
-    runs-on: ubuntu-latest
+  test:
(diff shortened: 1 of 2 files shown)


kvrigor (approved):

yguclu:
I have opened pyccel/psydac#475 in order to better test SymPDE with Python 3.13.

PR #184, opened 2025-12-10
empty Unions as empty lists

### LINKED ISSUE #183: iterating over an empty Union raises an error
an empty Union is a None object:
https://github.com/pyccel/sympde/blob/6fd903feea7fb0b308e7639d96a15b77a4edcc80/sympde/topology/basic.py#L139
which raises an error when trying to iterate over it. This is not convenient and should be fixed.

campospinto:
When a Union is created with no arguments, return an empty list instead of a None object. This should allow to iterate over empty "Union" objects. 

Fixes #183

### DIFF
--- a/sympde/topology/basic.py
+++ b/sympde/topology/basic.py
@@ -136,11 +136,11 @@ class Union(BasicDomain):
         # remove duplicates and sort domains by their string representation
         args = sorted(set(args), key=str)
 
-        # a. If the required Union contains no domains, return None;
+        # a. If the required Union contains no domains, return an empty list;
         # b. If it contains a single domain, return the domain itself;
         # c. If it contains multiple domains, create a Union object.
         if not args:
-            obj = None
+            obj = []
         elif len(args) == 1:
             obj = args[0]
         else:
--- a/sympde/topology/basic.py
+++ b/sympde/topology/basic.py
@@ -136,11 +136,11 @@ class Union(BasicDomain):
         # remove duplicates and sort domains by their string representation
         args = sorted(set(args), key=str)
 
-        # a. If the required Union contains no domains, return an empty list;
+        # a. If the required Union contains no domains, return an empty tuple;
         # b. If it contains a single domain, return the domain itself;
         # c. If it contains multiple domains, create a Union object.
         if not args:
-            obj = []
+            obj = ()
         elif len(args) == 1:
             obj = args[0]
         else:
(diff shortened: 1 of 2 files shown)


sympde/topology/basic.py
```diff
@@ -136,11 +136,11 @@ def __new__(cls, *args):
...
-        # a. If the required Union contains no domains, return None;
+        # a. If the required Union contains no domains, return an empty tuple;
         # b. If it contains a single domain, return the domain itself;
         # c. If it contains multiple domains, create a Union object.
         if not args:
-            obj = None
+            obj = ()
         elif len(args) == 1:
             obj = args[0]
         else:
```
yguclu, line 146: I am not sure I understand why we need this change for the case of an empty Union, but not for a Union containing a single domain. If being able to iterate is really important, why not doing it for all cases?

campospinto: Indeed, single domain unions should be handled consistently.
I just wanted to see whether the modification for empty unions was passing the tests before looking at the case of a single element.

I have no idea whether "unions" should be iterable objects, I just saw that we iterate on them in [some places](https://github.com/pyccel/psydac/blob/a89a6ee293f75cff099cc2d77e42b511ae91de51/psydac/feec/conforming_projectors.py#L801C14-L801C24) and currently this raises an error when the union is empty.

campospinto: ps -- according to the mathematical usage, a "union" of domains (sets in Rd) is a priori a domain itself, so it's unclear what we should get when iterating over it: subdomains ? points ? 

Maybe the issue comes from the fact that some objects expected to be iterable (like [interiors](https://github.com/pyccel/sympde/blob/6fd903feea7fb0b308e7639d96a15b77a4edcc80/sympde/topology/domain.py#L143), [boundaries](https://github.com/pyccel/sympde/blob/6fd903feea7fb0b308e7639d96a15b77a4edcc80/sympde/topology/domain.py#L152C9-L152C19) or [interfaces](https://github.com/pyccel/sympde/blob/6fd903feea7fb0b308e7639d96a15b77a4edcc80/sympde/topology/domain.py#L218C9-L218C19) ) are currently defined as "unions"...

campospinto: For now I'm turning this into a draft -- we should probably discuss in issue #183 how to better solve this problem

PR #187, opened 2026-05-03
Refactor `is_linear_expression()` in expr/expr.py

### LINKED ISSUE #121: Stuck or very Slow check is_linear_expression in LinearForm and BilinearForm
`is_linear_expression` checks whether the expressions given in a `LinearForm` (or a `BilinearForm` resp.ly) instance is actually linear (bi-linear resp.ly). For some expressions, known to be linear, the check is really slow or even stuck in a loop. This might be due to the current version used of sympy in sympde.
As an example, you may import `Laplacian` from `psydac.examples.poisson_2d_mapping` and `TargetMapping` from `sympde.topology.analytical_mapping` and run
```
params   = dict(c1 = 0, c2 = 0, k = 0.3, D = 0.2)
mapping  = TargetMapping('F', **params)
lapl  = Laplacian(mapping)
s, t  = mapping.logical_coordinates
x, y  = mapping.expressions
k     = params['k']
D     = params['D']
kx    = 2 * pi / (1 - k + D)
ky    = 2 * pi / (1 + k)
phi = (1 - s**8) * sin(kx * (x - 0.5)) * cos(ky * y)
rho = - lapl(phi)
logical_domain = Square('Omega', bounds1 = model.domain[0], bounds2 = model.domain[1])
domain  = mapping(logical_domain)
V0 = ScalarFunctionSpace('V0', domain)
u0, v0 = elements_of(V0, names = 'u0, v0')
rhs = LinearForm(v0, integral(domain, rho * v0))
```

alisa-kirkinskaia:
### 1. Fix #121:

Fix a bug in `is_linear_expression` related to `expand()` being very slow for some expressions. The bug is caused by the following line:

```python
if not( (a-b).expand() == 0 or a.expand() == b.expand()):
```

In the fix we avoid comparing the values of `a.expand()` and `b.expand()` if `(a-b).expand()` is 0, as it already means the expression is linear.

### 2. Refactor `is_linear_expression` for better readability:

- Remove unused argument `integral`
- Rename variables
- Add docstring 

### 3. Update library patch version to 0.19.3

### DIFF
--- a/sympde/expr/expr.py
+++ b/sympde/expr/expr.py
@@ -127,7 +127,7 @@ class LinearExpr(BasicExpr):
 
         args = _sanitize_arguments(arguments, is_linear=True)
 
-        if not is_linear_expression(expr, args, integral=False):
+        if not is_linear_expression(expr, args):
             msg = '> Expression is not linear'
             raise UnconsistentLinearExpressionError(msg)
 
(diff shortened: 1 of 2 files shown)


yguclu (changes requested):
Can you please update the library version to 0.19.3 in the file `sympde/pyproject.toml`?

sympde/expr/expr.py
```diff
@@ -739,71 +739,86 @@ def linearize(form, fields, trials=None):
...
+    """
+    Checks if expression is linear with respect to ``args``:
+
+        1. Additivity:   f(x + y) = f(x) + f(y)
+        2. Homogeneity:  f(alpha * x) = alpha * f(x)
+
+    Parameters
+    ----------
+    expr
+        Symbolic expression to test.
```
yguclu: I think `expr` should be of some basic SymPy type. Could you please try and find out which one? It would be nice to add this information to the docstring, and possibly do a type check, too.

sympde/expr/expr.py
```diff
@@ -739,71 +739,86 @@ def linearize(form, fields, trials=None):
...
+    args : iterable
+        Each argument must be ScalarFunction or VectorFunction.
+    debug : bool, optional
+        Print diagnostic info if a check fails.
+
+    Returns
+    -------
+    bool
+        True if the expression is linear with respect to all given arguments, False otherwise.
```
yguclu: We should check that `is_linear_expression` truly works for expressions with an arbitrary number of arguments. I am not sure if we do that already in any unit tests in folder `sympde/expr/`, could you please have a look?

sympde/expr/expr.py
```diff
@@ -739,71 +739,86 @@ def linearize(form, fields, trials=None):
...
+        x_args.append(x)
+        y_args.append(y)
+
+    # ---------------------------------------------------------------------------
+    # check addition property: f(x + y) = f(x) + f(y)
+    summed_args = [x + y for x, y in zip(x_args, y_args)]
+    expr_at_x  = expr.subs(zip(args, x_args)) # f(x)
+    expr_at_y = expr.subs(zip(args, y_args)) # f(y)
+    expr_at_sum    = expr.subs(zip(args, summed_args)) # f(x + y)
+    expected_sum = expr_at_x + expr_at_y # = f(x) + f (y)
```
yguclu: Minor formatting improvements:
```suggestion
    expr_at_sum = expr.subs(zip(args, summed_args)) # f(x + y)
    expected_sum = expr_at_x + expr_at_y # f(x) + f(y)
```

sympde/expr/expr.py
```diff
@@ -739,71 +739,86 @@ def linearize(form, fields, trials=None):
...
+
+    if (expr_at_sum - expected_sum).expand() != 0:
+        expr1 = expr_at_sum.expand()
+        expr2 = expected_sum.expand()
+        if expr1 != expr2:
+            if debug:
+                print('Failed to assert addition property')
+                print('{} != {}'.format(expr1, expr2))
```
yguclu: For long expressions, this error message will be quite difficult to read. Hence, I suggest expanding it across multiple lines:
```suggestion
                print(r"Failed to assert addition property `f(x + y) = f(x) + f(y)`, where:")
                print()
                print('f(x + y) =')
                print(expr1)
                print()
                print('f(x) + f(y) =')
                print(expr2)
```

Please note that I have used a Python [raw string](https://docs.python.org/3/library/re.html#raw-string-notation) in the first line of the suggested code, in order to avoid any issue with the markdown inline quote characters I wanted to print

sympde/expr/expr.py
```diff
@@ -739,71 +739,86 @@ def linearize(form, fields, trials=None):
...
+
+    scaled_expr = alpha * expr_at_x
+
+    if (expr_at_scaled_x - scaled_expr).expand() != 0:
+        expr1 = expr_at_scaled_x.expand()
+        expr2 = scaled_expr.expand()
+        if expr1 != expr2:
+            if debug:
+                print('Failed to assert multiplication property')
+                print('{} != {}'.format(expr1, expr2))
```
yguclu: Here I also suggest splitting the error message across multiple lines, as in my previous comment

sympde/expr/expr.py
```diff
@@ -739,71 +739,86 @@ def linearize(form, fields, trials=None):
...
+
+        1. Additivity:   f(x + y) = f(x) + f(y)
+        2. Homogeneity:  f(alpha * x) = alpha * f(x)
+
+    Parameters
+    ----------
+    expr
+        Symbolic expression to test.
+    args : iterable
+        Each argument must be ScalarFunction or VectorFunction.
```
yguclu, line 762: It is probably worth explaining in some detail that `x` and `y` at the top of this docstring correspond to different values of `args`, which is the list of arguments `expr` depends on

yguclu (changes requested):
Good job @alisa-kirkinskaia! I have a couple of minor comments, mostly about formatting.

yguclu:
> Can you please update the library version to 0.19.3 in the file `sympde/pyproject.toml`?

Thanks for commit 85fba96!

sympde/expr/expr.py
alisa-kirkinskaia: Done in [320ebaa](https://github.com/pyccel/sympde/commit/320ebaaf6af61d4b34c4341d662920ab531c2fee)

sympde/expr/expr.py
alisa-kirkinskaia: Done in [661c442](https://github.com/pyccel/sympde/commit/661c442b026bb2c6b96cf107451dc16c89f29ac0)

sympde/expr/expr.py
alisa-kirkinskaia: See previous comment

sympde/expr/expr.py
alisa-kirkinskaia: Done in [d87e74c](https://github.com/pyccel/sympde/commit/d87e74c4e8db2fbb9fa893fae92bee7d51c7fc1a)

sympde/expr/expr.py
alisa-kirkinskaia: It is done for expressions of 2 arguments, for example [here](https://github.com/pyccel/sympde/blob/master/sympde/expr/tests/test_expr.py#L63)

sympde/expr/expr.py
```diff
@@ -739,71 +739,103 @@ def linearize(form, fields, trials=None):
...
-    right_args = []
+def is_linear_expression(expr, args, debug=True):
+    """
+    Checks if expression is linear with respect to each argument in``args``:
+
+        1. Additivity:   f(x + y) = f(x) + f(y)
+        2. Homogeneity:  f(alpha * x) = alpha * f(x)
+
+    In this notation, x and y represent independent copies of the same argument. In general, f may have an
+    arbitrary number of arguments.
```
yguclu: Please keep the line lenght to 80 characters (79 + newline)
```suggestion
    Checks if expression is linear with respect to each argument in ``args``:

        1. Additivity:   f(x + y) = f(x) + f(y)
        2. Homogeneity:  f(alpha * x) = alpha * f(x)

    In general, `f` may have an arbitrary number of arguments: `x` and `y`
    represent independent copies of the same list of arguments.
    

```

sympde/expr/expr.py
yguclu: Thanks! Here, too, let's keep the line length to 80 characters

sympde/expr/expr.py
yguclu: Please keep the line length within 80 characters

yguclu (changes requested):

sympde/expr/expr.py
alisa-kirkinskaia: Does this rule only apply to comments/docstrings, or any line of code?

yguclu: To any line of (new) code, if possible. In the future, we can try and enforce it using a linter

alisa-kirkinskaia: OK. I changed it in [04040e1](https://github.com/pyccel/sympde/commit/04040e112ee032ca50b012320bcd6c79cb564267) and [f2c3505](https://github.com/pyccel/sympde/commit/f2c35056c80ad3decc162488e6f7a5eb5c6f8737) for all lines in `is_linear_expression`

yguclu: Thanks

yguclu (approved):
Looks good to me!

